PR #12, opened 2015-04-29
Introduced VxWorks Simulator root

RogerioDosSantos:
Introduced the root path for the VxWorks simulator.
It the "host:" is not handled, this can cause infinite recursive call on
VxWorks simulator when usinf the canonical function


RogerioDosSantos:
Because this pull request contains past changes I did, I am canceling it to have only the local change I want.

PR #13, opened 2015-04-29
Introduced VxWorks Simulator root and corrected canonical for VxWorks

RogerioDosSantos:
01 - Introduced the root path for the VxWorks simulator.
It the "host:" is not handled, this can cause infinite recursive call on
VxWorks simulator when using the canonical function

02 - Corrected path canonical function to handle properly path existence in
some platforms where the root separator might not be a valid path.
E.g.: On VxWorks the path "/" can be invalid but the path "/tgtsvr/" is
a valid path.

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -397,6 +397,7 @@ namespace
 //--------------------------------------------------------------------------------------//
 
   const char dot = '.';
+  const char* preferred_separator_string = "/";
 
   bool not_found_error(int errval)
   {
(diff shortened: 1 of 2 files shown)


RogerioDosSantos:
It is more than one year I requested this one. I am wondering when this will be evaluated and which additional steps should I do for those changes to be past of the boost release.

RogerioDosSantos:
03 years passed and not even a feedback on this item. What should I do for this item to be approved?

Lastique:
If this is still relevant, please, rebase against the current develop.

Also, I don't quite understand the change to `canonical`, although I didn't dig into the code deeply yet. What I'm concerned about is whether this change is needed specifically for VxWorks and noone else. If it's VxWorks-specific then it should be guarded with a platform macro check.

Lastique:
No activity, so considering this is not relevant.

PR #14, opened 2015-07-21
Let temp_directory_path return /data/local/tmp on Android

tiwoc:
See https://svn.boost.org/trac/boost/ticket/11491


Beman:
I'll hand apply this to make a stylistic change, and also to update the docs.

Thanks,

--Beman

PR #15, opened 2015-08-27
Fix a race condition in unique_path.

CornedBee:
If two threads call unique_path at the same time for the first time in the program run,
both initial calls to CryptAcquireContext can fail. Both threads will then call the function
with CRYPT_NEWKEYSET, but only one of these threads can succeed. The other will
fail with NTE_EXISTS.

This patch makes it so that if a call fails with that error, it will try to call without the
flag one more time, in case another thread created the key set in the meantime.

This also applies the patch from trac report #7506. Using these additional flags
is the right thing to do.

### DIFF
--- a/src/unique_path.cpp
+++ b/src/unique_path.cpp
@@ -41,6 +41,33 @@ void fail(int err, boost::system::error_code* ec)
   return;
 }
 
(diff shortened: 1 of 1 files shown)


Beman:
Thanks, Sebastian,

--Beman

PR #16, opened 2015-09-13
Remove unused deprecated include

MarcelRaad:
The comment in boost/iterator.hpp mentions that the file is obsolete and will be deprecated, and it is not used anyway.

Tested with MSVC 14.


Beman:
Thanks,

--Beman

PR #17, opened 2015-09-28
'some_function() noexcept = default;' fixes

apolukhin:
`some_function() noexcept = default;` seems like a bad practice. Clang3.6 will fail compilation if  exception specification of defaulted function does not match the calculated one. GCC4.8 and before also fail compilation. Latest versions of GCC silently ignore the explicit exception specification.

This pull request drops the `BOOST_NOEXCEPT` if followed by `= default`. This fixes the compilation errors like:

```
../../../boost/filesystem/operations.hpp:266:5: error: function ‘boost::filesystem::file_status::file_status(const boost::filesystem::file_status&)’ defaulted on its first declaration must not have an exception-specification

../../../boost/filesystem/operations.hpp:267:18: error: function ‘boost::filesystem::file_status& boost::filesystem::file_status::operator=(const boost::filesystem::file_status&)’ defaulted on its first declaration must not have an exception-specification

### DIFF
--- a/include/boost/filesystem/operations.hpp
+++ b/include/boost/filesystem/operations.hpp
@@ -263,14 +263,14 @@ namespace boost
                : m_value(v), m_perms(prms) {}
 
 # ifndef BOOST_NO_CXX11_DEFAULTED_FUNCTIONS
-    file_status(const file_status&) BOOST_NOEXCEPT = default;
-    file_status& operator=(const file_status&) BOOST_NOEXCEPT = default;
+    file_status(const file_status&) = default;
+    file_status& operator=(const file_status&) = default;
 # endif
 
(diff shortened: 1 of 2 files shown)


Beman:
Antony,

Thanks for the pull request. I fixed the problem in a slightly different way; I'd already wasted too much time with "= default" problems so eliminated its use for now.

Here is the commit message:

Fix pull request 17, 'some_function() noexcept = default;' from Antony Polukhin, by never relying on "= default". Antony's pull request would have fixed the problem interaction with noexcept by eliminating the noexcept. I preferred to retain the noexcept rather than the "= default", as applying "= default" has been exceedingly troublesome for both VC++ and GCC due to interactions between the two C++11 features. GCC interactions varied not just for the version, but also for the platform. 

Thanks,

--Beman

PR #22, opened 2015-12-16
Don't auto-link when the user requested that we don't

muggenhor:
Don't auto-link when the user requested that we don't


Beman:
I'm having trouble analyzing this. That's mostly my fault, as I wasn't even aware of the existence of BOOST_AUTO_LINK_NOMANGLE. If you could answer a couple of questions, that would help.

Why are you including <boost/filesystem/config.hpp>? It's already included by <boost/filesystem/operations.hpp>

Why isn't the code in <boost/filesystem/config.hpp> starting with the line "//  enable automatic library variant selection..." sufficient?

What steps are needed to reproduce the problem?

Thanks,

--Beman

muggenhor:
@Beman:
> Why are you including <boost/filesystem/config.hpp>? It's already included by <boost/filesystem/operations.hpp>

Mostly that's the include-what-you-use mentality at work here. I.e. I'm using `BOOST_FILESYTEM_NO_LIB` directly here, so including it directly.

> Why isn't the code in <boost/filesystem/config.hpp> starting with the line "// enable automatic library variant selection..." sufficient?

Because this isn't at all about variant selection, this change is about giving the library user control so that she can use _not_ to link to `advapi32.lib`.

> What steps are needed to reproduce the problem?

I see I didn't give a problem description here, my bad. In short there's two problems, one more of a clean design one, the other a practical one.

The design issue is that using auto-linking (in general) circumvents the build system, which results in hiding dependencies. This is a significant problem in the company I'm working because we're dealing with a lot of hidden/implicit dependencies and circular dependencies as a result.

The practical problem is that we're targeting WinCE, and on WinCE this library (advapi32) doesn't even exist. It's functionality does, it lives in another library (I can't recall which right now). But the way the code is currently written doesn't allow me to circumvent this problem in our build system, I have to go in and do source modifications. Then I have to keep these modifications up to date with every single Boost version upgrade we wish to do.

muggenhor:
Given that WinCE is no longer relevant for me anymore I'm going to leave this as is (i.e. not make a new PR).

PR #23, opened 2015-12-16
Pass function name as wide-char string on WinCE

muggenhor:
On Windows CE GetProcAddress takes its parameter as a wide-char string. Unfortunately no such equivalent (e.g. GetProcAddressW) exists on regular Windows. Hence the wrapper macro for its parameter.


muggenhor:
Updated the patch to take advantage of boostorg/winapi@27b1ac8139197a7feaf16097074e2a1c3e7628f1

PR #26, opened 2016-01-29
filesystem: getting file_type from a directory_entry without causing a system call on linux

taruti:
Trac https://svn.boost.org/trac/boost/ticket/11947

When iterating over a directory with directory_iterator the directory_entries are created with the file_type component of the m_symlink_status defined if the filesystem supports it on Linux due to BOOST_FILESYSTEM_STATUS_CACHE.

However the directory iteration does not produce the permission component of the file status. Thus if using somedirectoryentry.symlink_status().type() the library performs a superfluous lstat system call because status_known does not succeed due to permissions_present not succeeding.

This is possible to fix e.g. by adding a symlink_type method to the directory_entry class, a patch implementing it is linked below:

### DIFF
--- a/doc/reference.html
+++ b/doc/reference.html
@@ -2490,6 +2490,8 @@ and permissions of a file.</p>
         file_status  status(system::error_code&amp; ec) const;
         file_status  symlink_status() const;
         file_status  symlink_status(system::error_code&amp; ec) const;
+        file_type    symlink_type() const;
+        file_type    symlink_type(system::error_code& ec) const;
 
         bool operator&lt; (const directory_entry&amp; rhs);
         bool operator==(const directory_entry&amp; rhs); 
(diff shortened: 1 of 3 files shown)


Lastique:
You can already obtain this information by calling `symlink_status` (member of `directory_entry` or non-member) and then `type` on the returned `file_status`. I don't think a dedicated shortcut function is warranted here.

Lastique:
On the second thought, you are correct, I overlooked the permissions part. I might add the shortcut function then, but it should probably be universal (i.e. not specific to `directory_entry`). Also, the PR had conflicts, so I'll leave it closed.

taruti:
The PR was from 3 years ago, thus the conflicts. Ended up just using the C API for this few years ago.

PR #27, opened 2016-04-27
Rename generic() to generic_path(), since generic is a keyword in C++/CX

fxb:
When using Boost.Filesystem from a project compiled as C++/CX code,
compilation fails with a syntax error, because generic is a keyword.

    error C2059: syntax error: 'generic'

See section "Generic interfaces" in C++/CX here: https://msdn.microsoft.com/en-us/library/hh755792.aspx


fxb:
@Beman: I'm unsure if renaming this function is considered fine, because it will break the API. Since the comment preceding it says it is experimental, and the function is not included in the documentation yet, I went ahead and created this pull request. Let me know what you think.

Beman:
Thanks! Given that we already have a generic_string template, generic_path makes good sense as a name.
--Beman

PR #30, opened 2016-07-04
Fix warning for GCC under Cygwin

thtrummer:


Beman:
Applied. Thanks!

--Beman

PR #31, opened 2016-07-07
fix signature stem(), extension() member functions

faithandbrave:
`stem()` and `extension()` member functions doesn't have parameter.


Beman:
Thanks!

--Beman

PR #32, opened 2016-08-29
Fixed internal inconsistency from exception in end recursive dir iterator

ansingh:
If there’s an exception during increment, the check to make end
iterator is skipped. If this happens on the last element of recursive
directory iterator, the iterator is left in an inconsistent state.


Lastique:
This should be already fixed in the current develop.

PR #33, opened 2016-09-04
use the right directory_iterator contructor with the ec parameter

elfofo:
This fixes [7307](https://svn.boost.org/trac/boost/ticket/7307l) by setting the error code to "access denied".

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -385,7 +385,9 @@ namespace
 
     if (type == fs::directory_file)  // but not a directory symlink
     {
-      for (fs::directory_iterator itr(p);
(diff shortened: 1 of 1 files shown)


Beman:
Hopefully fixed. Here is the commit message: Fix #7307, remove_all(dirname,ec) throws on write protected directories. This is a tough one to test. There are three internal function calls where errors might arise, and it would take too much time to write tests for each of those cases. Someday we will have Titus Winter's mock installable file system, but for now are relying on code inspection rather than testing.

Thanks for the pull request! Although I didn't use it directly, it was helpful in coming up with the eventual fix.

--Beman

PR #35, opened 2016-11-09
Fix fs::create_directories(".") and fs::create_directories("..") internal assertion failure

Romain-Geissler-1A:
Hi,

I am having regressions between Boost 1.59 and 1.61 when calling fs::create_directories(".") or fs::create_directories(".."). It now fails with:
Assertion `(parent != p)&&("internal error: p == p.parent_path()")'

This is because "..".parent() is empty, and calling fs::create_directories("") is actually invalid.

Is this OK for the trunk ?

Cheers,
Romain

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -937,9 +937,9 @@ namespace detail
  BOOST_FILESYSTEM_DECL
   bool create_directories(const path& p, system::error_code* ec)
   {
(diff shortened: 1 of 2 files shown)


Beman:
Hi Romain,

This problem was also ticket #12495, and was fixed as a result of that ticket. The fix should be in time for the 1.63.0 release.

Thanks,

--Beman

PR #37, opened 2016-11-22
Ticket #12578: Make directory iterators able to detect when a copy has advanced to the end

mlimber:
Motivating example is using an iterator (e.g., testing to see if it is at the end) after a copy of it has been advanced to the end:
```cpp
auto di = fs::directory_iterator( "." );
const auto end = fs::directory_iterator();
const auto dist = std::distance( di, end ); // implicitly iterate a copy of di to the end
                                            // which makes di undetectably invalid.
std::cout << "File count: " << dist << std::endl;
while( di++ != end ) // crash! Also with: for( const auto& e : di ), etc.
{
  std::cout << *di << std::endl;
}
```
For details, see [Ticket #12578](https://svn.boost.org/trac/boost/ticket/12578).

### DIFF
--- a/include/boost/filesystem/operations.hpp
+++ b/include/boost/filesystem/operations.hpp
@@ -939,7 +939,12 @@ namespace detail
     void increment() { detail::directory_iterator_increment(*this, 0); }
 
     bool equal(const directory_iterator& rhs) const
-      { return m_imp == rhs.m_imp; }
+    {
+      const bool is_lhs_valid = m_imp && m_imp->handle;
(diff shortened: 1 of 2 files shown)


Beman:
I've applied essentially the same fix as your pull request, but with some added comments and a different set of test cases. By the way, the motivating example you give does a post-increment on an iterator than will be the end iterator the last time through the loop.

Thanks for the bug report!

--Beman

mlimber:
Fixed the increment. Thanks for processing it so quickly!

PR #39, opened 2016-12-11
Fix errno propagation in space(p, ec).

cmuellner:
The space(p, ec) implementation assigned a wrong
error value to the given error_code object.
Instead of getting the error value from errno,
the code used the value '-1!=0'.

Signed-off-by: Christoph Müllner <christophm30@gmail.com>


cmuellner:
A real-life problem, which is hit by the not propagated error value, can be found here:
https://github.com/facebook/hhvm/pull/7534

Beman:
Thanks!

--Beman

PR #41, opened 2017-03-14
Patch for ticket #10731 and ticket #9480: Evaluate path.extension onl…

Dani-Hub:
…y once, perform (case-insensitive) string comparisons using the wchar_t variants from the CRT to prevent code conversions via path.string() calls.

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -516,26 +516,140 @@ namespace
       || errval == ERROR_BAD_PATHNAME  // "//nosuch" on Win64
       || errval == ERROR_BAD_NETPATH;  // "//nosuch" on Win32
   }
+  
+  // File name case-insensitive comparison needs to be locale- and collation-independent. 
+  // The approach used below follows a combined strategy described here
+  // http://archives.miloush.net/michkap/archive/2005/10/17/481600.html
+  // and here:
+  // http://archives.miloush.net/michkap/archive/2007/09/14/4900107.html
+  // If possible, we use CompareStringOrdinal, alternatively we fallback to 
+  // RtlEqualUnicodeString, and if all fails we perform the equivalent characterwise 
+  // comparsion using LCMapString and uppercase binary equality.
 
-// some distributions of mingw as early as GLIBCXX__ 20110325 have _wcsicmp, but the
-// offical 4.6.2 release with __GLIBCXX__ 20111026  doesn't. Play it safe for now, and
-// only use _wcsicmp if _MSC_VER is defined
-#if defined(_MSC_VER) // || (defined(__GLIBCXX__) && __GLIBCXX__ >= 20110325)
-#  define BOOST_FILESYSTEM_STRICMP _wcsicmp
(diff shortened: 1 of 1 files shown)


sergey-shambir:
We also faced with the same problem: unnecessary code convertions cause exceptions for several valid paths, inluding Cyrillic and Chinese paths. This PR fixes wrong exceptions.

Dani-Hub:
I should emphasize that the suggested PR uses the same level of case-sensitivity comparison correctness as the existing code. I investigated the matter more deeply and it seems that a better file-name case-sensitity comparison should use CompareStringOrdinal (From Windows Vista on) and LCMapString + binary comparison (For older Windows version), respectively. If interest exists, I'm willing to improve my current PR following that direction.

Dani-Hub:
I investigated more and found that RtlInitUnicodeString did had significant performance impact on RtlEqualUnicodeString. The following table shows an extended experiment using the same conditions as described in the previous comment:

_wcsicmp (zeroterm.):
 1.282223s wall, 1.281250s user + 0.000000s system = 1.281250s CPU (99.9%)
LCMapStringW (characterwise):
 7.970858s wall, 7.968750s user + 0.000000s system = 7.968750s CPU (100.0%)
LCMapStringW + wmemcmp (chunks):
 6.186876s wall, 6.171875s user + 0.000000s system = 6.171875s CPU (99.8%)
CompareStringOrdinal (zeroterm.):
 1.061088s wall, 1.046875s user + 0.000000s system = 1.046875s CPU (98.7%)
RtlEqualUnicodeString + RtlInitUnicodeString (zeroterm.):
 2.586483s wall, 2.562500s user + 0.000000s system = 2.562500s CPU (99.1%)
RtlEqualUnicodeString + wcslen (zeroterm.):
 0.873219s wall, 0.875000s user + 0.000000s system = 0.875000s CPU (100.2%)
CompareStringOrdinal (+length):
 0.815105s wall, 0.828125s user + 0.000000s system = 0.828125s CPU (101.6%)
RtlEqualUnicodeString (+length):
 0.876963s wall, 0.875000s user + 0.000000s system = 0.875000s CPU (99.8%) 

It turns out that CompareStringOrdinal is just a wrapper for RtlCompareUnicodeString, see

http://archives.miloush.net/michkap/archive/2007/10/12/5396685.html

so in the end I decided to simplify the previous resolution by ignoring CompareStringOrdinal and simply using RtlEqualUnicodeString + wcslen when RtlEqualUnicodeString is found, falling back to LCMapString (which should rarely ever occur, but since that is a run-time test, we need a fallback). This is what the following commit suggests.

Dani-Hub:
Any chance for considering this patch suggestion for Boost 1.64? If my advanced suggestion looks too far-reaching to you, I would be totally happy with the initial Patch ID 755766a.

Beman:
Thanks, Daniel!

Fix applied to develop. Will merge to master after tests have cycled.

--Beman

PR #42, opened 2017-03-22
noexcept is_empty throws on missing permissions.

ldqrk:
`boost::filesystem::is_emtpy` with `boost::system::error_code` throws on missing permissions.


Beman:
The editor used for the P/R removed trailing spaces from lines. That causes unwanted churn. So I made the changes by hand.

Thanks for the P/R!

--Beman

PR #45, opened 2017-06-05
Fix read_symlink with windows junctions points

edgare:
On windows `is_symlink` considers both symbolic links and junction points to be symlinks. However `read_symlink` assumes a symbolic sink, which results in reading from bad offsets in the `REPARSE_DATA_BUFFER` struct.


Lastique:
Should be fixed by https://github.com/boostorg/filesystem/pull/100.

PR #49, opened 2017-07-26
Update PGI C++ compiler support

dkolsen-pgi:
Remove an #if from operations.cpp that used to be necessary for the file to compile with PGI C++, but causes compilation errors with PGI C++ compilers of the last few years.

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -2143,10 +2143,6 @@ namespace
     return ok;
   }
 
-#if defined(__PGI) && defined(__USE_FILE_OFFSET64)
-#define dirent dirent64
-#endif
-
   error_code dir_itr_first(void *& handle, void *& buffer,
     const char* dir, string& target,
     fs::file_status &, fs::file_status &)


Lastique:
If the workaround is still needed for older versions, why not guard it with a version check?

dkolsen-pgi:
There is no version check because we don't know of any customers who are still using a PGI compiler old enough to require that workaround.  If there are any such users, there's a good chance that they also have an older version of Boost that won't have this change.  In my opinion, the potential benefit of a version check is smaller than the cost of cluttering up the code.

alexeikh:
(I am not Boost project member, nor have commit rights to Boost, just trying to contribute with the code review.)

I think, this pull request is nice and @dkolsen-pgi 's reasoning make sense. Old workarounds clutter the code and are liabilities. They should have finite lifetime. This workaround was committed in 2011 (commit d776d3985fcf76572375bb0a885f250d014b6ea1). Time to retire it, IMO.

The pull request comes from an employee of the company which produces the PGI compiler, IIUC. Which gives the PR more credibility.

I am voting, for what it's worth, for this PR to be merged in.

PR #52, opened 2017-09-04
operations: Don't set _FILE_OFFSET_BITS for older Android versions

tiwoc:
The build was broken when building for Android < 7.0 with the latest
Android NDK (r15c). With earlier NDKs, the build succeeded, but silently
used 32b offsets.


KayEss:
I'd just forked the project to make a very similar change. Would be really nice to see this merged

Lastique:
See also https://github.com/boostorg/filesystem/pull/67.

alexeikh:
Now when #67 has been merged, this pull request is not need any more and can be closed, I think.

pdimov:
Leave a comment

PR #53, opened 2017-09-19
Fix remove_all(path, ec) to not throw filesystem_error via ++itr

b-spencer:
This pull request is in relation to [Boost Trac 13217](https://svn.boost.org/trac10/ticket/13217).

According to the documentation, uintmax_t remove_all(const path& p,
system::error_code& ec) is not supposed to throw filesystem_error. However,
when the directory_iterator used in the loop inside the implementation helper
function remove_all_aux() is incremented, it's not done in a way that
considers the ec argument. So, when incrementing the iterator fails, the
operator++() on the directory_iterator throws, which leads to remove_all()
incorrectly throwing.

To fix this, it seems that one can take the same approach as the recent
changes for Fix #7307
https://github.com/boostorg/filesystem/commit/4e4374336c640c2717f6a1b168406e11bdf7d6f1
and replace the ++itr call in the loop with an explicit call to either
directory_iterator::increment(system::error_code& ec) whenever an ec argument
is in effect. This should prevent remove_all() from throwing in such cases and
thus allow it to conform to its interface specification.

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -2406,7 +2406,7 @@ namespace detail
         && (filename.size()== 1
           || (filename[1] == dot
             && filename.size()== 2)))
-        {  it.increment(*ec); }
+        {  ec ? it.increment(*ec) : ++it; }
     }
   }
 


b-spencer:
I did not search for other places where operator++() is used on directory iterator when no exceptions should be thrown.

b-spencer:
I found a related issue in `directory_iterator_constructor` and added a fix for that.

Beman:
Will look at this sometime next week.

Thanks,

--Beman

b-spencer:
Is there anything I can do to help?

b-spencer:
@pdimov @Beman: Boost 1.66.0 was released back in December and seems to still lacks this and other fixes.  What are the obstacles to merging this PR?

pdimov:
Something like https://github.com/boostorg/filesystem/commit/350109a99164463caf60af1a90090cce15ef75ae?

pdimov:
I'm not sure that the returned `count` is correct when the iteration errors out though; the initial value of 1 is presumably because it includes `p` but we never get to removing it.

b-spencer:
Yes, 350109a is almost exactly what I was going to change this to.

I agree that `count` looks wrong in the directory case.  It seems like `count` should start as `0` and then be incremented just _after_ the `remove_file_or_directory()` statement, unless `ec && *ec`.

b-spencer:
But, note the return value of the [standardized version](https://en.cppreference.com/w/cpp/filesystem/remove).  It always returns `uintmax(-1)` on error in the non-throwing case.  Perhaps this should just do the same (and say so in the docs).

pdimov:
Returning -1 instead of number removed until the error seems to throw away information for no reason, so I won't be switching to that.

b-spencer:
@pdimov, do you need anything else from me here?  I think you've taken over, which works for me.

pdimov:
Should be fixed in develop by https://github.com/boostorg/filesystem/commit/350109a99164463caf60af1a90090cce15ef75ae and https://github.com/boostorg/filesystem/commit/e6df54ce81bf78b7034a11cad74735723be4fe4a.

PR #44, opened 2017-04-27
fix compiler warnings about narrowing conversions

DanielaE:
Signed-off-by: Daniela Engert <dani@ngrt.de>

PR #54, opened 2017-10-05
Cleanup: Silence Visual Studio 2017 warnings.

KevinHopps:


MarcelRaad:
I've also been bugged by those warnings. There's already #44 from April.

Lastique:
Looks like you introduced tabs with your edits. Please, convert to spaces.

jeking3:
@KevinHopps would you be able to convert the tabs to spaces in your pull request to align with the current code style?  Thanks.

src/unique_path.cpp
```diff
@@ -111,7 +111,9 @@ void system_crypt_random(void* buf, std::size_t len, boost::system::error_code*
 
   if (!errval)
   {
-    BOOL gen_ok = ::CryptGenRandom(handle, len, static_cast<unsigned char*>(buf));
+    // Use static cast to silence Visual Studio 2017
+    // warning C4267: 'argument': conversion from 'size_t' to 'DWORD', possible loss of data
+    BOOL gen_ok = ::CryptGenRandom(handle, static_cast<DWORD>(len), static_cast<unsigned char*>(buf));
```
jeking3: @Lastique here's another random implementation that is not ready for Windows UWP.

jeking3: Should we be using `static_cast` or should we be using `boost::numeric_cast` to ensure proper handling of math here (and the other instances I see in the other file).

KevinHopps: numeric_cast seems appropriate. I will make that change.

alexeikh:
Now in 2 of 3 places the comment says "Use static cast", but the code uses `numeric_cast`.

jeking3:
In general, the comments in the code here are unnecessary.

src/windows_file_codecvt.cpp
```diff
@@ -59,8 +60,10 @@
     UINT codepage = AreFileApisANSI() ? CP_ACP : CP_OEMCP;
 
     int count;
+    // Use static cast to silence Visual Studio 2017:
+    // warning C4244: 'argument': conversion from '__int64' to 'int', possible loss of data
     if ((count = ::WideCharToMultiByte(codepage, WC_NO_BEST_FIT_CHARS, from,
-      from_end - from, to, to_end - to, 0, 0)) == 0)
+       boost::numeric_cast<int>(from_end - from), to, boost::numeric_cast<int>(to_end - to), 0, 0)) == 0)
```
alexeikh: Sorry for nitpicking again. This line has 1 extra indentation space, compared to the original line.

Also, this line has become very long. I'd rewrite the whole fragment, for example, as:

```c++
int count = ::WideCharToMultiByte(
  codepage, WC_NO_BEST_FIT_CHARS,
  from, boost::numeric_cast<int>(from_end - from),
  to, boost::numeric_cast<int>(to_end - to),
  0, 0);
if (count == 0)
  return error;  // conversion failed
```

src/windows_file_codecvt.cpp
```diff
@@ -40,7 +41,7 @@
 
     int count;
     if ((count = ::MultiByteToWideChar(codepage, MB_PRECOMPOSED, from,
-      from_end - from, to, to_end - to)) == 0) 
+      boost::numeric_cast<int>(from_end - from), to, boost::numeric_cast<int>(to_end - to))) == 0)
     {
       return error;  // conversion failed
     }
```
alexeikh: The expression in `if` is now very long. Suggesting to rewrite as:

```c++
int count = ::MultiByteToWideChar(
  codepage, MB_PRECOMPOSED,
  from, boost::numeric_cast<int>(from_end - from),
  to, boost::numeric_cast<int>(to_end - to));
if (count == 0)
  return error;  // conversion failed
```

Lastique:
Seems to be already fixed by d88945564a4ed037474f9a5952b87f4f9244d4ba.

PR #55, opened 2017-10-18
some directory_entry overloads for some status query functions

emmett-b:
Without these functions, calling (for example) `is_directory` with a `directory_entry` results in:
- a conversion to `path` using `operator const boost::filesystem::path&()`,
- then a call to `is_directory(const path& p)` which recomputes the status with `detail::status(p)`.
  
These functions avoid a costly recomputation of the status if one calls `is_directory(e)` instead of `is_directory(e.status)`

### DIFF
--- a/include/boost/filesystem/operations.hpp
+++ b/include/boost/filesystem/operations.hpp
@@ -830,6 +830,28 @@ private:
   file_status m_get_symlink_status(system::error_code* ec=0) const;
 }; // directory_entry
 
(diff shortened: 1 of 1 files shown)


Lastique:
Thanks. Committed 4731ce79de213ed9c75cfc3b125a0626efd57f07 with some modifications.

emmett-b:
Great! Thanks!

PR #59, opened 2017-11-20
Initialize function pointers at first use (II)

Dani-Hub:
Fixes static initialization issue of new compare function using interlocked functions. Kudus to Daniela Engert for her initial pull request ff8c321a655490783c7aac589796a5ee35a749e1.

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -96,6 +96,7 @@ using std::wstring;
 
 # else // BOOST_WINDOW_API
 
+#   include <boost/detail/interlocked.hpp>
 #   if (defined(__MINGW32__) || defined(__CYGWIN__)) && !defined(WINVER)
       // Versions of MinGW or Cygwin that support Filesystem V3 support at least WINVER 0x501.
       // See MinGW's windef.h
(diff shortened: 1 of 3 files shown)


src/operations.cpp
```diff
@@ -614,8 +613,51 @@ namespace
...
+    return rtl_equal_unicode_string_api ? equal_string_ordinal_ic_1
+                                        : equal_string_ordinal_ic_2;
+  }
+
+  static Ptr_equal_string_ordinal_ic compare_function;
+  static long compare_function_state;
+
+  inline Ptr_equal_string_ordinal_ic compare()
+  {
+    for (unsigned i = 0; compare_function_state < 2; ++i)
```
Lastique: This reads `compare_function_state` non-atomically (i.e. not thread-safe).

Dani-Hub: Ouch, you are completely right! I thought that I could use the `<atomic>`-free interlocked approach that `boost/interprocess/detail/win32_api.hpp` uses, but now I realize that the corresponding code in the `function_address_holder::get_module` and `function_address_holder::get` is defect as well! I will take a look this night or tomorrow to improve the patch - thanks for the review so far!

Dani-Hub: The non-atomical read access should have been fixed now via https://github.com/Dani-Hub/filesystem/commit/76c061ec79328cc4fee6a23f07e2918c66fe44a7

Dani-Hub: I have now created ticket https://svn.boost.org/trac10/ticket/13309 to handle the same problem in `boost/interprocess/detail/win32_api.hpp` (twice)

PR #62, opened 2017-11-22
Use ad-hoc extension comparison to avoid initialization issues

pdimov:
Alternative to #43 and #59.

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -518,115 +518,28 @@ namespace
       || errval == ERROR_BAD_PATHNAME  // "//nosuch" on Win64
       || errval == ERROR_BAD_NETPATH;  // "//nosuch" on Win32
   }
-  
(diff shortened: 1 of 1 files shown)


DanielaE:
Simple as it is, it is fine with me. 👍

pdimov:
Would also supersede #50.

DanielaE:
This problem seems to have been discovered over and over again 😄

Dani-Hub:
I vote for your revised patch and withdraw my further revised patch https://github.com/Dani-Hub/filesystem/commit/76c061ec79328cc4fee6a23f07e2918c66fe44a7 and https://github.com/Dani-Hub/filesystem/commit/63d758789fa102a9be1741522d6f5d232e39ce4e. If the future shows a proper need for my icase correct code, we can restore and refine it further. For now I keep it in my private code where I have practical uses for it.

pdimov:
Thanks Daniel. Just need @Beman's blessing then.

Issue #58, opened 2017-10-30
recursive_directory_iterator noexcept constructor may call std::terminate


vamireh:
[recursive_directory_iterator constructor](https://github.com/boostorg/filesystem/blob/develop/include/boost/filesystem/operations.hpp#L1173) marked as noexcept but `new detail::recur_dir_itr_imp` may throw std::bad_alloc

PR #63, opened 2017-11-22
Do not use throwing new in noexcept constructors

pdimov:
This resolves issue #58.

### DIFF
--- a/include/boost/filesystem/operations.hpp
+++ b/include/boost/filesystem/operations.hpp
@@ -40,6 +40,7 @@
 #include <ctime>
 #include <vector>
 #include <stack>
+#include <new> // std::nothrow
 
 #ifdef BOOST_WINDOWS_API
 #  include <fstream>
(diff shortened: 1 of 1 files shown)


include/boost/filesystem/operations.hpp
```diff
@@ -1173,22 +1174,42 @@ namespace filesystem
...
+      if (pi == 0)
+      {
+        ec = make_error_code(boost::system::errc::not_enough_memory);
+      }
+      else
+      {
+        m_imp.reset(pi);
+
+        m_imp->m_options = opt;
+        m_imp->m_stack.push(directory_iterator(dir_path, ec));
```
Lastique: This `push` can still throw, doesn't it? Maybe it's better to wrap the constructor and initialization list in `try`/`catch` instead? The same with the other constructor.

jeking3:
@pdimov this needs a follow-up from you.

PR #64, opened 2017-12-01
fix NULL pointer dereference (reproduced with increased inline-depth in clang)

proller:


b-spencer:
Also fixed in https://github.com/boostorg/filesystem/pull/53

pdimov:
Fixed by #77.

Issue #65, opened 2017-12-07
Compile error for Android Level API lower then 21 with NDK r16


Bjoe:
Boost 1.65.1
Compiler: clang from android NDK
clang_version__ "5.0.300080 "
NDK version: 16.0.4442984
Build system: cmake (3.6.4111459)
Std librariy: libc++_static
Build machine: Ubuntu 16.04.3 LTS

Since Google [introduced unified headers](https://android.googlesource.com/platform/ndk/+/master/docs/UnifiedHeaders.md) the following compile error occurs to build boost::filesystem for example Android Level 19:
```
clang-linux.compile.c++.without-pth bin.v2/libs/filesystem/build/clang-linux-android/release/link-static/target-os-android/threading-multi/operations.o
libs/filesystem/src/operations.cpp:1705:12: error: no member named 'truncate' in the global namespace
    error(!BOOST_RESIZE_FILE(p.c_str(), size) ? BOOST_ERRNO : 0, p, ec,
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
libs/filesystem/src/operations.cpp:224:38: note: expanded from macro 'BOOST_RESIZE_FILE'
#   define BOOST_RESIZE_FILE(P,SZ)(::truncate(P, SZ)== 0)
                                   ~~^
1 error generated.

[...]

PR #67, opened 2018-03-01
Don't define _FILE_OFFSET_BITS=64 on Google Android NDK up to API level 21

Lastique:
The newer Google NDK versions don't declare truncate() function when
`_FILE_OFFSET_BITS=64` is defined. Don't define that macro and fall back to 32-bit
offsets.

Should fix https://github.com/boostorg/filesystem/issues/65

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -10,7 +10,9 @@
 
 //--------------------------------------------------------------------------------------// 
 
-//  define 64-bit offset macros BEFORE including boost/config.hpp (see ticket #5355) 
+//  define 64-bit offset macros BEFORE including boost/config.hpp (see ticket #5355)
+//  Google Android NDK is broken though as it doesn't provide truncate() declaration when _FILE_OFFSET_BITS=64 is defined (https://github.com/boostorg/filesystem/issues/65)
+#if !defined(__ANDROID__) || (__ANDROID_API__+0) >= 21 || defined(__CRYSTAX__)
 #if !(defined(__HP_aCC) && defined(_ILP32) && !defined(_STATVFS_ACPP_PROBLEMS_FIXED))
 #define _FILE_OFFSET_BITS 64 // at worst, these defines may have no effect,
 #endif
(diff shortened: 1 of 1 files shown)


Bjoe:
Hi. @Lastique looks like my patch file [android-boost-1_65_1-patches.patch.txt](https://github.com/boostorg/filesystem/files/1538797/android-boost-1_65_1-patches.patch.txt) and as I explained [on 7 Dec 2017 (see first message)](https://github.com/boostorg/filesystem/issues/65#issue-280092873) before google add the "unified headers solution", it was silently ignore the define of _FILE_OFFSET_BITS=64 in lower API level then 21 and uses truncate() instate of truncate64(). This is really worst! Now we can see at compile time that feature/function xy like truncate64() is not supported by this API level. This is really an improvement! So I think this statement "...NDK is broken though as it doesn't provide truncate() ..." is not a fairly comment ;-) ... its not supporetd in lower API levels...
Anyways... the real question and thats why I didn't create a pull request, it fixes the compile issue but it is not something I will add in the production code or? What happens if boost::filesystem::resize_file() should resize a large file with truncate() instate of truncate64() on android? Maybe the filesystem is getting corrupted...
I think the correct solution is to improve the BOOST_RESIZE_FILE macro and not silently ignore the fact that the android NDK lower then API 21 still not supported truncate64(). But as I written [on 7 Dec 2017 (see first message)](https://github.com/boostorg/filesystem/issues/65#issue-280092873) ... I still waiting for some contributors.

Lastique:
I don't see how one can resize a file to more than 2 GiB on a system that does not support 64-bit file offsets.

Regarding the "broken" comment, I understand the rationale behind the current behavior of NDK, but I still don't think it is correct. NDK should really not support `_FILE_OFFSET_BITS=64` if it can't provide the full support for it. I will change the comment, though, if the library maintainer asks.

Lastique:
I've added a check for whether the size argument to `resize_file` fits in the `off_t` value.

rcdailey:
I verified this fixes my specific issue over on #65. Can we get this merged and put into a hotfix for 1.66?

Lastique:
Closing in favor of https://github.com/boostorg/filesystem/pull/69.

PR #70, opened 2018-03-11
Prevent crash if calling filesystem::copy without an error parameter.

jorj1988:


amerry:
See also #71, which looks like a cleaner fix for the same issue.

pdimov:
Fixed by #71.

PR #71, opened 2018-04-06
Fix undefined behavior in detail::copy

rcombs:
Call the detail versions, which take pointers, rather than converting the pointers to references and back (which is UB).

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -890,20 +890,20 @@ namespace detail
   BOOST_FILESYSTEM_DECL
   void copy(const path& from, const path& to, system::error_code* ec)
   {
-    file_status s(symlink_status(from, *ec));
+    file_status s(detail::symlink_status(from, ec));
     if (ec != 0 && *ec) return;
 
     if(is_symlink(s))
     {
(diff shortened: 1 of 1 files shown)


grembo:
Could you please merge this? This is causing segfaults in production. Thanks!

rcombs:
Yeah I should've been more clear, this causes crashes.

grembo:
The code segment below demonstrates the undefined behaviour (in my case leading to segfaults):

```
  int main()
  {
    boost::filesystem::copy("/tmp/a", "/tmp/b");
  }
```

See here for a more detailed description of the problem (adding the patch to the freebsd ports tree): https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=229888

pokey909:
@Beman could you please take a look and merge this patch if there are no concerns with the code?

jeking3:
Perhaps the CI suite should be running UBSAN and static code analyses?

pdimov:
It is running UBSAN.

jeking3:
Apologies, I looked at the build results for the PR which is 7 months old.

PR #74, opened 2018-06-08
Correct handling of status() for reparse point when it is not a symlink.

stima:
When the path is Windows Reparse Point and it is not a symlink return status as reparse_file, otherwise resolve symlink and return status as regural_file with attributes of resolved file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -270,7 +270,9 @@ matrix:
       env: TOOLSET=clang COMPILER=clang++ CXXSTD=03,11,14,1z
 
 install:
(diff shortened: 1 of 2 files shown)


stima:
Hello guys, is the any changes to have it reviewed and committed?

stima:
Hi guys, any updates?

Flamefire:
As I'm working on something related: What exactly is the problem you fixed here? Do you have a testcase that fails? Can you add that to the tests?

stima:
Hi. I will create test soon.

PR #75, opened 2018-06-12
Create u8path and u8string function

ken2812221:
u8path and u8string are C++17 functions, but they are not in boost. So I create it


Empact:
How about a test?

PR #78, opened 2018-07-14
path.hpp: Unify coding style of path_traits::convert() calls

datadiode:


Empact:
These are all relative to `begin()`, and dereferencing `begin()` is valid so long as the string is not `empty()`. Guards are in place here, so it seems the existing code is fine.

datadiode:
Empact, you are right. I don't know why vc140 gave me an assertion in `path::string()`. I drew a wrong conclusion which falsely implied that `+` evaluated before `&*`, which would have lead to dereferencing `begin() + size()`. Using `c_str()` instead of `begin()` might still be considered preferable as it is less tricky, and in accordance with all the other calls to `path_traits::convert()` which exist in the same file.

PR #92, opened 2018-11-22
Fixed polymorphic object slicing warning

Flast:
see: https://wandbox.org/permlink/IXitHfg9USwGkhll


Lastique:
Note that I had to revert my change since apparently slicing was intentional. See https://github.com/boostorg/filesystem/commit/5a70ced692d36bf30996d521b35a68f7011ef53b.

Flast:
Oops... Sorry for annoying.

PR #100, opened 2019-02-03
Fix canonical and read_symlink for junction points

Flamefire:
Resolving junction points in read_symlink and canonical is wrong making it unusable: It silently returns the wrong results.

Issues identified:
- read_symlink does not handle junction points and reads the wrong struct
- canonical tries to resolve the root_name as a symlink. On windows this could be `C:` and refers to the current working directory on C. Hence depending if the current dir is a symlink canonical resolves the root_name to the current dir which is wrong

Supersedes #45

- [x] Includes #142

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -792,6 +792,12 @@ path canonical(const path& p, const path& base, system::error_code* ec)
 
       result /= *itr;
 
+      // If we don't have an absolute path yet then don't check symlink status.
+      // This avoids checking "C:" which is "the current directory on drive C"
+      // and hence not what we want to check/resolve here.
+      if(!result.is_absolute())
+          continue;
+
       bool is_sym (is_symlink(detail::symlink_status(result, ec)));
       if (ec && *ec)
         return path();
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -1546,12 +1546,30 @@ path read_symlink(const path& p, system::error_code* ec)
   if (!error(::DeviceIoControl(h.handle, FSCTL_GET_REPARSE_POINT,
         0, 0, info.buf, sizeof(info), &sz, 0) == 0 ? BOOST_ERRNO : 0, p, ec,
         "boost::filesystem::read_symlink" ))
+  {
+    const wchar_t* buffer;
+    size_t offset, len;
+    if (info.rdb.ReparseTag == IO_REPARSE_TAG_MOUNT_POINT)
+    {
+      buffer = info.rdb.MountPointReparseBuffer.PathBuffer;
+      offset = info.rdb.MountPointReparseBuffer.PrintNameOffset;
+      len = info.rdb.MountPointReparseBuffer.PrintNameLength;
+    } else if (info.rdb.ReparseTag == IO_REPARSE_TAG_SYMLINK)
+    {
+      buffer = info.rdb.SymbolicLinkReparseBuffer.PathBuffer;
+      offset = info.rdb.SymbolicLinkReparseBuffer.PrintNameOffset;
+      len = info.rdb.SymbolicLinkReparseBuffer.PrintNameLength;
+      // Note: iff info.rdb.SymbolicLinkReparseBuffer.Flags & SYMLINK_FLAG_RELATIVE
+      //       -> resulting path is relative to the source
+    } else
+    {
+      error(BOOST_ERROR_NOT_SUPPORTED, p, ec, "boost::filesystem::read_symlink");
+      return symlink_path;
+    }
     symlink_path.assign(
-      static_cast<wchar_t*>(info.rdb.SymbolicLinkReparseBuffer.PathBuffer)
-      + info.rdb.SymbolicLinkReparseBuffer.PrintNameOffset/sizeof(wchar_t),
-      static_cast<wchar_t*>(info.rdb.SymbolicLinkReparseBuffer.PathBuffer)
-      + info.rdb.SymbolicLinkReparseBuffer.PrintNameOffset/sizeof(wchar_t)
-      + info.rdb.SymbolicLinkReparseBuffer.PrintNameLength/sizeof(wchar_t));
+      buffer + offset / sizeof(wchar_t),
+      buffer + (offset + len) / sizeof(wchar_t));
+  }
 # endif
   return symlink_path;
 }
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -1569,7 +1569,7 @@ path read_symlink(const path& p, system::error_code* ec)
       //       -> resulting path is relative to the source
     } else
     {
-      error(BOOST_ERROR_NOT_SUPPORTED, p, ec, "boost::filesystem::read_symlink");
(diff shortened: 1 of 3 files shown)


Flamefire:
Test was verified to trigger the bug: https://ci.appveyor.com/project/pdimov/filesystem/builds/22102228/job/0k3ueppx1l562jm5

Flamefire:
There is a test failure remaining on MinGW resulting from a to low or missing define for `_WIN32_WINNT`. What shall be done about this? IMO it should be defined to the value for WinXP at the least if it is not defined and error if it is defined to low. Even WinXP is out-of-support so it can be dropped especially for new boost versions.

Lastique:
Boost doesn't drop Windows versions, it's Windows SDKs that may drop some Windows versions. Legacy MinGW is pretty conservative, it appears to have stopped evolving before it had any reasonable support for NT6 APIs, so XP is a good choice for that target. With regard to Boost.Filesystem, I'm not sure there is any NT6 API used in the library, so it may make sense to target XP on any Windows SDK.

stima:
@Flamefire I added tests for my pull request https://github.com/boostorg/filesystem/pull/102.

mloskot:
Is this PR related to issues and patches submitted here? 
https://svn.boost.org/trac10/ticket/10900

Flamefire:
@mloskot It seems it adresses the same issue. Can't really compare the patches though. If this change is wanted, I can rebase

GitEvgen:
How do I find out if this has been released and in which version?
Somewhat related to this issue: https://stackoverflow.com/questions/44393241/boostfilesystemrelative-cannot-access-the-file-because-it-is-being-used-by#

Flamefire:
@mloskot @pdimov Requesting review of this PR. Rebased to current develop after the latest changes (indentation and such) caused major conflicts.

Not sure why it has been pending for so long given that on Windows a wrong struct is read which could be a security vulnerability. But also forgot about this before the 1.73 release

Flamefire:
@mloskot Thanks, that's a good idea. Wasn't that familiar with B2 back then. And syntax of it is hard too, needed a colon, not comma ;-)

Flamefire:
Also fixed a failed conflict resolution. Should be good now. Let's see what CI says. I'm no expert of this either but I'm confident in the changes. Especially as it now doesn't blindly reinterpret_cast a pointer but checks the "enum" first and fails for unknown values.

mloskot:
> needed a colon, not comma ;-)

Right, I forgot the `,` is a conjunction e.g. `<target-os>windows,<toolset>gcc:<build>no`

Flamefire:
@mloskot @Lastique Finally fixed the remaining issues with the test. Factored out some stuff into #142 to keep this PR focused on the bugfix however this is rebased on #142 so that should be merged first.

src/operations.cpp
```diff
@@ -1546,12 +1552,30 @@ path read_symlink(const path& p, system::error_code* ec)
   if (!error(::DeviceIoControl(h.handle, FSCTL_GET_REPARSE_POINT,
         0, 0, info.buf, sizeof(info), &sz, 0) == 0 ? BOOST_ERRNO : 0, p, ec,
         "boost::filesystem::read_symlink" ))
+  {
+    const wchar_t* buffer;
+    size_t offset, len;
```
Lastique, line 1557: `std::size_t`

src/operations.cpp
```diff
@@ -792,6 +792,12 @@ path canonical(const path& p, const path& base, system::error_code* ec)
 
       result /= *itr;
 
+      // If we don't have an absolute path yet then don't check symlink status.
+      // This avoids checking "C:" which is "the current directory on drive C"
+      // and hence not what we want to check/resolve here.
+      if(!result.is_absolute())
```
Lastique, line 798: (nitpick) Add space after `if`.

src/operations.cpp
```diff
@@ -1546,12 +1552,30 @@ path read_symlink(const path& p, system::error_code* ec)
...
         "boost::filesystem::read_symlink" ))
+  {
+    const wchar_t* buffer;
+    size_t offset, len;
+    if (info.rdb.ReparseTag == IO_REPARSE_TAG_MOUNT_POINT)
+    {
+      buffer = info.rdb.MountPointReparseBuffer.PathBuffer;
+      offset = info.rdb.MountPointReparseBuffer.PrintNameOffset;
+      len = info.rdb.MountPointReparseBuffer.PrintNameLength;
+    } else if (info.rdb.ReparseTag == IO_REPARSE_TAG_SYMLINK)
```
Lastique, line 1563: (nitpick) Add newline before `else`. Ditto in the following code.

Or even better, convert to a `switch`/`case` block.

test/issues/99_canonical_with_junction_point.cpp
```diff
@@ -0,0 +1,61 @@
+
```
Lastique, line 1: License header missing.

Also, please add a short description of what this test is testing.

test/issues/99_canonical_with_junction_point.cpp
```diff
@@ -0,0 +1,61 @@
...
+  ~TmpDir()
+  {
+    boost::system::error_code ec;
+    fs::remove_all(path, ec);
+  }
+};
+
+int main()
+{
+  if(std::system("mklink /?") != 0)
```
Lastique, line 26: (nitpick) Add a space after `if`.

src/operations.cpp
Flamefire: Good point! 👍

test/issues/99_canonical_with_junction_point.cpp
```diff
@@ -0,0 +1,70 @@
...
+    boost::system::error_code ec;
+    fs::remove_all(path, ec);
+  }
+};
+
+// Test fs::canonical for various path in a Windows directory junction point
+// This failed before due to broken handling of absolute paths and ignored ReparseTag
+int main()
+{
+  if (std::system("mklink /?") != 0)
```
Lastique, line 35: This check doesn't work on Windows 8.1. `mklink` exists, it prints help, and then the test decides it doesn't exist and exits.

Lastique: I don't know why `std::system` returns 1, but it does. It also returns 1 if the file doesn't exist. I think `std::system` result is not indicative and not trustworthy, we need to find another way. Possibly, from the Jamfile.

GitEvgen: Maybe try running it and check if the link got created. Only skip if it wasn't. 
Also the test should include `mklink /D ...` The failures were different with that type of link. And there could be differences whether it's an absolute or relative link to absolute or relative path.

Lastique: I've already implemented a different check in the Jamfile.

Lastique: Another problem with this test is that it doesn't remove the temporary directory it creates. `remove_all` fails at removing "<temp_dir>/real/sub" with error 32 (directory used by another process).

PR #103, opened 2019-02-07
Replacing static_casts with numeric_casts for narrowing.

KevinHopps:


Lastique:
I don't think this is correct, at least not everywhere. There are places where an exception must not be thrown (when `error_code` is used to report errors).

Also, I'm not fond of introducing a dependency on Boost.NumericConversion in Boost.Filesystem. If we are going to check for overflows, I'd like it to be done locally.

KevinHopps:
@Lastique, I agree. Exceptions should not be thrown from a function that reports errors through an error code.

PR #106, opened 2019-03-08
[CMake] Add minimal cmake support

Mike-Devel:
Functionality:

- CMake file only supports add_subdirectory workflow.
- Compiles boost filesystem (without name mangling).
- Provides target Boost::filesystem to link against.
- Does NOT compile/run unit tests (yet) and
  doesn't support installation.

For basic testing:

1) Make sure you have checked out the latest (develop) version of all other boost libraries you are depending on (directly or indirectly) into sibling folders of this library (e.g. recursively check out boostorg/boost). 

2) Copy a file like this: https://github.com/Mike-Devel/boost/blob/boost_cmake_demo/CMakeLists.txt into the boost root directory. Essentially, that file will simply call `add_subdirectory( )` on all boost libraries with a cmake file to make sure all dependencies are known to cmake. However, it will also ignore some libraries that are known to be troublesome with the add_subdirectory workflow.

3) Then (from boost root) 

         mkdir __build__ && cd __build__
         cmake ..
         cmake --build libs/filesystem

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,3 +47,8 @@ target_link_libraries( boost_filesystem
         Boost::system
         Boost::type_traits
 )
+
+target_link_libraries( boost_filesystem
+    PRIVATE
+        Boost::winapi
+)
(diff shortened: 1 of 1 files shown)


Lastique:
Dependency on Boost.WinAPI is missing.

Mike-Devel:
Thanks for spotting this. Must have forgotten the `--track-sources` option when checking with boostdep

Lastique:
Thanks.

PR #107, opened 2019-04-02
operations_test: Enable locale query on MinGW/Cygwin

Kojoley:
The `GetUserDefaultUILanguage` is available for a quite long time already (about 19 years for Cygwin, and 7-12 years for MinGW).

Fixes test fails on non-EN system locales. More info on the symmetrical PR for System https://github.com/boostorg/system/pull/40


Lastique:
Thanks.

PR #109, opened 2019-04-16
Fixed to check readdir_r errno

semun-lee:
readdir_r() returns errno when it fails.
errno should be set properly.

Change-Id: Ia79c74bf73ebd53174ba37cb8afd2abda49c47e5
Signed-off-by: Semun Lee <semun.lee@samsung.com>


Lastique:
Should be fixed in 28ccca4.

Issue #110, opened 2019-05-20
const reference on temporary returned by boost::filesystem::path::operator/ is broken


MarcoWagner:
After upgrading from boost 1.69 to boost 1.70 holding a const reference from a temporary returned by ``boost::filesystem::path::operator/`` is broken.

boost-Version: 1.70.0
Platform: Windows 10 / Version 1803 (Build 17134.765)
Compiler: Visual Studio 2017 (cl.exe, 19.16.27030.1; link.exe, 14.16.27030)

[main.cpp.txt](https://github.com/boostorg/filesystem/files/3197136/main.cpp.txt)
[CMakeLists.txt](https://github.com/boostorg/filesystem/files/3197154/CMakeLists.txt)

PR #111, opened 2019-06-19
path: fix dangling reference return

CJBussey:
* issue #110


Lastique:
Thanks.

Lastique:
```
path const& p = path("abc") / "def"
```

`p` is left a dangling reference because although const references do bind to rvalue references, they don't extend the lifetime of the referred object in this case.

rcdailey:
So it's because of the unnamed temporaries. Makes sense. Thank you.

PR #126, opened 2019-12-17
Include the public quoted header from IO

glenfe:
<boost/io/detail/quoted_manip.hpp> is now just a deprecated forwarding header for backward compatibility, and will be removed eventually.


Flamefire:
It already is deprecated and the build throws a warning.

PR #134, opened 2020-04-11
Add creation_time API

### LINKED ISSUE #133: Add creation_time API in operations.hpp
ITNOA

I think it is good to add API to provide creation time of path. with below signature

```
std::time_t creation_time(const path& p)

std::time_t creation_time(const path& p, system::error_code& ec) BOOST_NOEXCEPT
```

soroshsabz:
ITNOA

This PR resolved #133 

I added support for below platform

- [x] windows
- [x] linux with glibc greater than 2.28
- [x]  Linux with glibc less than 2.28
- [x] FreeBSD
- [x] Mac OSX greater than Mac OS X 10.5 (Leopard)

I add another platform in coming days.

I would be happy to review my PR.

thanks a lot


test/operations_test.cpp
```diff
@@ -1194,7 +1194,7 @@ namespace
     ec.clear();
     BOOST_TEST(!fs::create_directories("/foo", ec));  // may be OK on Windows
                                                       //  but unlikely to be OK on POSIX
-    BOOST_TEST(ec);
+    BOOST_TEST(!ec);
```
Lastique, line 1197: Why this change?

src/operations.cpp
```diff
@@ -1399,6 +1399,40 @@ void last_write_time(const path& p, const std::time_t new_time,
 # endif
 }
 
+BOOST_FILESYSTEM_DECL
+std::time_t creation_time(const path& p, system::error_code* ec)
+{
+# ifdef BOOST_POSIX_API
+
+  struct statx path_stat;
+  if (error(::statx(AT_FDCWD, p.c_str(), AT_SYMLINK_NOFOLLOW | AT_NO_AUTOMOUNT |
```
Lastique, line 1408: The presence of `statx` needs to be tested at compile time.

test/operations_test.cpp
soroshsabz: I revert it.

src/operations.cpp
soroshsabz: I check it right now :)

soroshsabz:
Finally I can make green Linux test in travis :) the only one build remaining in travis, it is Mac OSX

soroshsabz:
@Lastique Thanks for review my code I try to resolve all comments and support all desired platforms. so I happy to review it again.

Thanks

soroshsabz:
@Lastique I need this feature, so I thankful for review it :)

src/linux_statx.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+//  See library home page at http://www.boost.org/libs/filesystem
+
+//--------------------------------------------------------------------------------------//
+
+#ifndef BOOST_FILESYSTEM3_SRC_LINUX_STATX_HPP_
+#define BOOST_FILESYSTEM3_SRC_LINUX_STATX_HPP_
+
+// Only for internal use
+// Copy many contents from stat.h
```
Lastique, line 18: I don't like this. Please, don't reimplement libc. Aside from maintenance cost, it is also unclear from the licensing perspective (you cannot copy/paste third party code and relicense it under BSL). If libc does not implement the necessary API then make Boost.Filesystem return `ENOSYS`.

src/operations.cpp
```diff
@@ -119,6 +128,17 @@
 #   if _POSIX_C_SOURCE < 200809L
 #     include <utime.h>
 #   endif
+#   ifdef __linux__
+#     ifdef __GLIBC__
+#       include <gnu/libc-version.h>
```
Lastique, line 133: No internal header includes, please. glibc version macros are not defined in this header anyway. Use `unistd.h` or another standard header to get them.

src/operations.cpp
```diff
@@ -119,6 +128,17 @@
...
+#       if __GLIBC_MINOR__ > 27
```
Lastique, line 134: Use `__GLIBC_PREREQ` to test glibc version.

test/operations_test.cpp
```diff
@@ -253,6 +257,11 @@ namespace
     fs::path to;
   };
 
+  void bad_creation_time()
+  {
+    fs::creation_time(".");
```
Lastique, line 262: Why is this guaranteed to throw? Is this a portable assumption? I could imagine, it should return the creation time of the current directory.

If you want to simulate error, try passing a path to a non-existing file, for example.

src/operations.cpp
```diff
@@ -119,6 +128,17 @@
...
+#         define BOOST_HAS_NATIVE_STATX 1
+#       endif
+#     endif
+#     ifndef BOOST_HAS_NATIVE_STATX
+#       include "linux_statx.hpp"
```
Lastique, line 139: Remove this, as per the comment for this header.

src/operations.cpp
```diff
@@ -94,9 +94,18 @@
 #endif
 
 # ifdef BOOST_POSIX_API
-
 #   include <sys/types.h>
-#   include <sys/stat.h>
+#   if defined(__APPLE__)
+// The stat64() system calls first appeared in Mac OS X 10.5 (Leopard)
+#     if (defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && __MAC_OS_X_VERSION_MIN_REQUIRED >= 1060)
+#       define _DARWIN_USE_64_BIT_INODE
```
Lastique, line 101: This macro must be defined before any headers are included. Any header may potentially include `sys/stat.h`, directly or indirectly.

src/operations.cpp
```diff
@@ -1399,6 +1419,67 @@ void last_write_time(const path& p, const std::time_t new_time,
...
+  struct statx path_stat;
+  if (error(::statx(AT_FDCWD, p.c_str(), AT_SYMLINK_NOFOLLOW | AT_NO_AUTOMOUNT |
+	AT_STATX_SYNC_AS_STAT, STATX_BTIME, &path_stat)!= 0 ? BOOST_ERRNO : 0,
+    p, ec, "boost::filesystem::creation_time"))
+      return std::time_t(-1);
+  return std::time_t(path_stat.stx_btime.tv_sec);
+# elif defined(__APPLE__)
+  // The stat64() system calls first appeared in Mac OS X 10.5 (Leopard)
+#   if (defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && __MAC_OS_X_VERSION_MIN_REQUIRED >= 1060)
+#     ifdef _DARWIN_USE_64_BIT_INODE
```
Lastique, line 1436: This is incorrect, as this is the macro you define to request the 64-bit inodes. Whether the feature is enabled is indicated by `_DARWIN_FEATURE_64_BIT_INODE`. http://polarhome.com/service/man/?qf=stat&tf=2&of=Darwin&sf=2

src/operations.cpp
```diff
@@ -1399,6 +1419,67 @@ void last_write_time(const path& p, const std::time_t new_time,
...
+# ifdef __linux__
+  struct statx path_stat;
+  if (error(::statx(AT_FDCWD, p.c_str(), AT_SYMLINK_NOFOLLOW | AT_NO_AUTOMOUNT |
+	AT_STATX_SYNC_AS_STAT, STATX_BTIME, &path_stat)!= 0 ? BOOST_ERRNO : 0,
+    p, ec, "boost::filesystem::creation_time"))
+      return std::time_t(-1);
+  return std::time_t(path_stat.stx_btime.tv_sec);
+# elif defined(__APPLE__)
+  // The stat64() system calls first appeared in Mac OS X 10.5 (Leopard)
+#   if (defined(__MAC_OS_X_VERSION_MIN_REQUIRED) && __MAC_OS_X_VERSION_MIN_REQUIRED >= 1060)
```
Lastique, line 1435: I may be missing something, but it looks like when this condition is not satisfied, the function is left empty.

If `st_birthtimespec` is not available on older versions of Mac OS X then the function must return `ENOSYS`.

src/operations.cpp
```diff
@@ -1399,6 +1419,67 @@ void last_write_time(const path& p, const std::time_t new_time,
...
+      return std::time_t(-1);
+  return std::time_t(path_stat.st_birthtimespec.tv_sec);
+#     endif
+#   endif
+# elif defined(__FreeBSD__)
+  struct stat path_stat;
+  if (error(::stat(p.c_str(), &path_stat) != 0 ? BOOST_ERRNO : 0, p, ec,
+	"boost::filesystem::creation_time"))
+      return std::time_t(-1);
+  return std::time_t(path_stat.st_birthtim.tv_sec);
```
Lastique, line 1455: I think, `st_birthtime` is a more portable way to access this field instead of `st_birthtim.tv_sec`. Also, NetBSD seems to support it as well: https://netbsd.gw.com/cgi-bin/man-cgi?stat+2.i386+NetBSD-8.0

test/operations_test.cpp
```diff
@@ -49,6 +49,10 @@ using std::endl;
 #include <ctime>
 #include <cstdlib> // for system(), getenv(), etc.
 
+#if defined(BOOST_POSIX_API) && defined(__linux__)
+# include <linux/version.h>
```
Lastique, line 53: Is this necessary? Could you run the tests unconditionally and simply skip them if you get `ENOSYS`?

test/operations_test.cpp
```diff
@@ -1754,6 +1763,26 @@ namespace
...
+# ifdef BOOST_POSIX_API
+#  ifdef __linux__
+#    if LINUX_VERSION_CODE < KERNEL_VERSION(4,11,0)
+    BOOST_TEST(CHECK_EXCEPTION(bad_creation_time, ENOSYS));
+#    endif
+#  endif
+# else
+    std::time_t ft = fs::last_write_time(f1x);
+    BOOST_TEST(ft == fs::creation_time(f1x));
+# endif
```
Lastique, line 1783: This doesn't run the test on POSIX systems. Just run the tests unconditionally and return if you get `ENOSYS`.

src/operations.cpp
```diff
@@ -1399,6 +1419,67 @@ void last_write_time(const path& p, const std::time_t new_time,
...
+# else
+  // TODO: Check for symbolic file
+  error(BOOST_ERRNO, p, ec, "boost::filesystem::creation_time does not support on this platform");
```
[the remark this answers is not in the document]
soroshsabz: I correct it

src/linux_statx.hpp
soroshsabz: @Lastique  I think it is very useful to implement internal implementation of statx syscall when glibc does not supported and it is cause to our library is more useful, but as you wish I remove it.

test/operations_test.cpp
soroshsabz: I fix it.

src/operations.cpp
soroshsabz: it is only work, when _POSIX_SOURCE not defined, so I think it is dangerous to use this.

Please see https://www.freebsd.org/cgi/man.cgi?query=stat&sektion=2

soroshsabz:
@Lastique I am sorry for my long delay, but I think I resolves many of your comment, and you can review it again.

thanks a lot

Lastique:
Thanks for your work. I have looked through your code and there were issues with it. For example, you are not checking the `stx_mask` field to see if the requested data was actually retrieved. Also, `AT_SYMLINK_NOFOLLOW` is not correct, as the function has to resolve symlinks, like `last_write_time` does. There were other issues as well, so I decided to implement this myself, taking some bits of your PR as an example. My current version is [here](https://github.com/boostorg/filesystem/commit/1eec6f443d1fb1c17a05822caa10eba87635f897).

PR #142, opened 2020-04-25
Remove API level checks in symlink functions in favor of runtime checks

Flamefire:
#100 became big and mixed unrelated things. So here is just the part making the symlink functions available if the runtime supports it. This is done by checking whether the functions exist (as before) and the compile time checks are not required. Only issue seen is `SYMBOLIC_LINK_FLAG_DIRECTORY` being not defined on low API levels, but as other implementations do: Just define it to the documented value for those rare cases (e.g. Cygwin using a pre XP default API level)

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -1529,7 +1529,7 @@ path read_symlink(const path& p, system::error_code* ec)
 
   union info_t
   {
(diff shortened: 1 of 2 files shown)


Flamefire:
@Lastique Ready :)

PR #144, opened 2020-05-23
Add WASI platform support

whitequark:
[WASI](https://wasi.dev/) is a POSIX-like platform that has a few important differences:
  * There is no concept of a mutable current directory. However, WASI host environment can provide a directory called `.` in the filesystem namespace that fulfills a similar role.
  * There is no concept of a file mode, and no `chmod` call.
  * There is no concept of a filesystem, and no `statfs` call.
  * There is no `readdir_r` call because there are no threads.

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -99,7 +99,9 @@
 
 #   include <sys/types.h>
 #   include <sys/stat.h>
-#   if !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__ANDROID__) && !defined(__VXWORKS__)
+#   if defined(__wasm)
+// WASI does not have statfs or statvfs.
+#   elif !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__ANDROID__) && !defined(__VXWORKS__)
 #     include <sys/statvfs.h>
 #     define BOOST_STATVFS statvfs
 #     define BOOST_STATVFS_F_FRSIZE vfs.f_frsize
@@ -1564,7 +1566,11 @@ void create_symlink(const path& to, const path& from, error_code* ec)
 BOOST_FILESYSTEM_DECL
 path current_path(error_code* ec)
 {
-# ifdef BOOST_POSIX_API
+# if defined(__wasm)
+  error(BOOST_ERROR_NOT_SUPPORTED, ec,
+    "boost::filesystem::current_path");
+  return path();
+# elif defined(BOOST_POSIX_API)
   struct local
   {
     static bool getcwd_error(error_code* ec)
@@ -1632,7 +1638,7 @@ path current_path(error_code* ec)
 BOOST_FILESYSTEM_DECL
 void current_path(const path& p, system::error_code* ec)
 {
-# ifdef UNDER_CE
+# if defined(UNDER_CE) || defined(__wasm)
   error(BOOST_ERROR_NOT_SUPPORTED, p, ec,
     "boost::filesystem::current_path");
 # else
@@ -1932,7 +1938,10 @@ void permissions(const path& p, perms prms, system::error_code* ec)
   if ((prms & add_perms) && (prms & remove_perms))  // precondition failed
     return;
 
-# ifdef BOOST_POSIX_API
+# if defined(__wasm)
+  error(BOOST_ERROR_NOT_SUPPORTED, p, ec,
+    "boost::filesystem::permissions");
+# elif defined(BOOST_POSIX_API)
   error_code local_ec;
   file_status current_status((prms & symlink_perms)
                              ? fs::symlink_status(p, local_ec)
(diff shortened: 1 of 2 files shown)


src/operations.cpp
```diff
@@ -99,7 +99,8 @@
 
 #   include <sys/types.h>
 #   include <sys/stat.h>
-#   if !defined(__APPLE__) && !defined(__OpenBSD__) && !defined(__ANDROID__) && !defined(__VXWORKS__)
+#   if defined(__wasm)
```
marcusmueller, line 102: This would be nicer than the existing code base, but I'd honestly think that future readers would enjoy a comment here, especially if WASI should ever be extended to cover things like `stat`.

whitequark: You're right, of course--I lost the comment while porting the diff to `develop` from 1.72.0. Fixed now.

whitequark: Note that this is specifically `statfs` and `statvfs`, not `stat` (it does have `stat`). `statfs` gets filesystem statistics, not file statistics.

src/operations.cpp
```diff
@@ -2186,6 +2193,12 @@ space_info space(const path& p, error_code* ec)
   if (ec)
     ec->clear();
 
+# if defined(__wasm)
+
+  if (!error(BOOST_ERROR_NOT_SUPPORTED,
+    p, ec, "boost::filesystem::space"))
+  {}
+
 # ifdef BOOST_POSIX_API
```
nabijaczleweli, line 2202: `#elif defined(BOOST_POSIX_API)` here as well?

src/operations.cpp
```diff
@@ -1564,7 +1566,9 @@ void create_symlink(const path& to, const path& from, error_code* ec)
 BOOST_FILESYSTEM_DECL
 path current_path(error_code* ec)
 {
-# ifdef BOOST_POSIX_API
+# if defined(__wasm)
+  return ".";
```
Lastique, line 1570: This is not right, `current_path` must return an absolute path (see C++20 [fs.op.current.path]). For example, `copy` implementation relies on that, and probably not only that.

If the platform does not support `current_path`, it should fail with `BOOST_ERROR_NOT_SUPPORTED` (`ENOSYS`), but then I'm not sure how useful the rest of the library would be as it is used in implementation of quite a few operations.

Lastique: As an alternative, you could assume that the current path is always the root directory. I'm not sure how valid is that in WASI (i.e. is `equivalent(".", "/") == true`).

src/operations.cpp
whitequark: Thanks, good catch

src/operations.cpp
Lastique: Also, see if there is a way to obtain path from a file descriptor or from the result of `stat` on WASI.

whitequark: Understood. I'll check whether this assumption holds (the precise WASI semantics around `.` are not entirely clear to me) and see what can be done.

whitequark: > Also, see if there is a way to obtain path from a file descriptor or from the result of `stat` on WASI.

That's what the [`fd_prestat_dir_name` syscall](https://github.com/WebAssembly/WASI/blob/master/phases/snapshot/docs.md#-fd_prestat_dir_namefd-fd-path-pointeru8-path_len-size---errno) is for.

Lastique: Can you use it to obtain the full path to '.'?

whitequark: > If the platform does not support `current_path`, it should fail with `BOOST_ERROR_NOT_SUPPORTED` (`ENOSYS`), but then I'm not sure how useful the rest of the library would be as it is used in implementation of quite a few operations.

I've changed the code to do that. It turns out that the application I'm porting is only using `boost::filesystem::path`, so it's still useful for that. Failing with `BOOST_ERROR_NOT_SUPPORTED` seems to be a safe choice that can be always revisited later, perhaps with input from WASI folks.

whitequark: > Can you use it to obtain the full path to '.'?

No. There is no concept of an "absolute path" in WASI, since it eliminates ambient authority and uses capabilities for everything; the `/<something>` directory, `.` directory, and `..` directory behave effectively like separate mounts, akin to the way `.` and `..` work on FAT, if you've seen that. It's kind of weird.

src/operations.cpp
```diff
@@ -1564,7 +1566,11 @@ void create_symlink(const path& to, const path& from, error_code* ec)
 BOOST_FILESYSTEM_DECL
 path current_path(error_code* ec)
 {
-# ifdef BOOST_POSIX_API
+# if defined(__wasm)
+  error(BOOST_ERROR_NOT_SUPPORTED, ec,
+    "boost::filesystem::current_path");
+  return "";
```
Lastique, line 1572: `return path()` please.

whitequark: Done

src/operations.cpp
```diff
@@ -2186,7 +2195,13 @@ space_info space(const path& p, error_code* ec)
   if (ec)
     ec->clear();
 
-# ifdef BOOST_POSIX_API
+# if defined(__wasm)
+
+  if (!error(BOOST_ERROR_NOT_SUPPORTED,
```
Lastique, line 2200: Use `emit_error` here, with no condition.

whitequark: Done

Lastique:
Thanks.

whitequark:
Thanks for the quick review!

PR #145, opened 2020-05-29
Use BOOST_OVERRIDE

EugeneZelenko:
Use BOOST_OVERRIDE to fix GCC -Wsuggest-override and Clang-tidy modernize-use-override warnings.
Fix Clang -Wextra-semi and Clang-tidy readability-container-size-empty and readability-redundant-control-flow warnings.
Alphabetical sort of STL headers.

### DIFF
--- a/include/boost/filesystem/exception.hpp
+++ b/include/boost/filesystem/exception.hpp
@@ -58,7 +58,7 @@ public:
   filesystem_error(filesystem_error const& that);
   filesystem_error& operator= (filesystem_error const& that);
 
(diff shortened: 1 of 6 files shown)


Lastique:
I'm not going to add `override` to destructors, this markup is pointless. Removing `return`s is also pointless, but ok. Replacing `size` with `empty` is useful. I don't see the need to reorder includes; if there is a reason, please describe it.

This PR is doing many things at once. Please extract the useful stuff in separate PRs.

Lastique:
> I don't see the need to reorder includes; if there is a reason, please describe it.

Sorry, I missed that you ordered them in alphabetical order. No, I'm not doing that.

PR #148, opened 2020-06-04
is_symlink(directory_entry) should use symlink_status().

kanje:
This fixes a problem that is_symlink(directory_entry) always returns
false, even if directory_entry is indeed a symlink. This change makes
is_symlink(directory_entry) behave the same as is_symlink(path) and
use symlink_status().


Lastique:
Thanks.

PR #150, opened 2020-07-02
Restore enabling Large File Support when building directory tools

ala-ableton:
See commit message for more details.

It might make sense to extract the 43 lines that are then duplicated between `src/directory.cpp` and `src/operations.cpp` to a separate file, but I'll let the maintainers decide.


Lastique:
I've made a larger fix for this. Thanks for the report.

ala-ableton:
@Lastique thanks for the quick fix!

PR #154, opened 2020-07-08
doc: Remove dead link to id complete_note

MarcoFalke:
No idea whether the note has been removed by accident or on purpose. If it has been removed by accident, the note should be re-added somewhere.

Also, fix a typo.


Lastique:
Thanks for the report. The `complete` operation is deprecated and its docs are no longer present in the reference. I've updated the docs to refer to `absolute` operation instead.

PR #163, opened 2020-10-16
Return default locale on POSIX systems

nre-ableton:
The comment here is incorrect; an empty locale is not the correct
behavior here and will cause exceptions to be thrown in other code.
Instead a locale object with the default constructor should be
returned, not one created with an empty string. This is a known issue
which was originally reported here:

https://svn.boost.org/trac10/ticket/4688

The issue claims to have been fixed in 7bb19f9 (see also:
https://svn.boost.org/trac10/changeset/72855).

However, this only fixes the issue for FreeBSD and not other POSIX
platforms. This patch is based on the one originally submitted here:

https://svn.boost.org/trac10/attachment/ticket/4688/boost_filesystem.patch

### DIFF
--- a/src/path.cpp
+++ b/src/path.cpp
@@ -887,9 +887,8 @@ namespace
     std::locale global_loc = std::locale();
     return std::locale(global_loc, new boost::filesystem::detail::utf8_codecvt_facet);
 # else  // Other POSIX
-    // ISO C calls std::locale("") "the locale-specific native environment", and this
-    // locale is the default for many POSIX-based operating systems such as Linux.
-    return std::locale("");
+    // Return a default locale object.
+    return std::locale();
 # endif
   }
 


nre-ableton:
See also: https://github.com/boostorg/process/pull/179

Lastique:
The C++ standard explicitly [lists](http://eel.is/c++draft/locale.cons#4) "" as one of the valid locale names. The tests run without errors on Linux systems both locally and the CI. Are you having a problem on a specific system? What system and what compiler?

nre-ableton:
@Lastique it's unfortunately been awhile since I have worked on this code, so I don't remember what the exact issue was. As close as I can remember, it was not a compiler error but a runtime error with the application, which crashed when it was launched.

Lastique:
I think the current code is correct. Default-constructed `std::locale` object is not necessarily the environment locale. It is created as a copy of the locale set by `std::locale::global()`, which by default is the same as the environment locale, but can be overridden by the user. The empty locale name explicitly requests the environment locale, not global. Creating such a locale must succeed. If a standard library rejects it then this is either a bug in the standard library or somehow weirdly misconfigured system locale.

If the problem reproduces, a patch that works around the bug for a particular C++ standard library and target system might be useful. I'll emphasize that this *may* be acceptable only as a workaround for a particular C++ standard library bug.

nre-ableton:
@Lastique fair enough, thanks for looking at this patch!

PR #169, opened 2020-12-20
Update WASI platform support

whitequark:
This is essentially a part of #144 that I've missed because I tested against an outdated version of Boost at that time. See that PR for details.

### DIFF
--- a/src/operations.cpp
+++ b/src/operations.cpp
@@ -1237,9 +1237,13 @@ bool copy_file(const path& from, const path& to, unsigned int options, error_cod
     goto fail;
   }
 
+#if defined(__wasm)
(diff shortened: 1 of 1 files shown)


Lastique:
If there is no access mode in WASI, then I'd prefer to also disable adding `S_IWUSR` above, along with its associated comment.

whitequark:
Something like this?

Lastique:
More like this:

```
mode_t to_mode = from_mode;
#if !defined(__wasm)
...
#endif
```

Lastique:
Merged to develop.

whitequark:
Thanks!

PR #177, opened 2021-03-12
Fix misplaced link opening tag for is_directory

hyarion:
The opening tag was placed between the letter i and s in is_directory instead of before the i.
This PR fixes that small mistake.


Lastique:
Thanks.

PR #241, opened 2022-07-07
Ignore unused variables

alandefreitas:


Lastique:
Sorry, but no, I'm not willing to maintain workarounds like this. If the warnings are bothering you, please disable it with a pragma or a compiler option.

Lastique:
Actually, it's even disabled [here](https://github.com/boostorg/filesystem/blob/fcc11010a5899d6a16d3f0a9d60704f88d62eada/include/boost/filesystem/detail/header.hpp#L39). It's just `header.hpp`/`footer.hpp` aren't included in the .cpp files.

PR #267, opened 2022-12-01
Fallback to file_directory_information_format for directory iterator

### LINKED ISSUE #266: directory_iterator does not work in some cases
directory_iterator does not work in Azure App Service environment when created over any folder in C:\home.
That environment is special with home folder mounted there in some way.

It fails with error 87 in:
file_id_extd_dir_info_format
file_full_dir_info_format
file_id_both_dir_info_format

I debuged it, and looks like it works in default case (file_directory_information_format), but it never fallback to file_directory_information_format, so logic in default case is never executed.

In the same environment current code works if iterator created over some other folders.

Problem reproducible on at least boost 1.79 and 1.80. Potentially same issue with #261 and #255.

iskiselev:
Ugly patch that fixed #266 for me

### DIFF
--- a/src/directory.cpp
+++ b/src/directory.cpp
@@ -879,6 +879,13 @@ error_code dir_itr_create(boost::intrusive_ptr< detail::dir_itr_imp >& imp, fs::
             {
                 DWORD error = ::GetLastError();
 
+                if (error == ERROR_NOT_SUPPORTED || error == ERROR_INVALID_PARAMETER)
+                {
+                    // Fall back to file_directory_information_format
+                    filesystem::detail::atomic_store_relaxed(g_extra_data_format, file_directory_information_format);
+                    return dir_itr_create(imp, dir, opts, params, first_filename, sf, symlink_sf);
+                }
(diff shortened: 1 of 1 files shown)


src/directory.cpp
```diff
@@ -879,6 +879,13 @@ error_code dir_itr_create(boost::intrusive_ptr< detail::dir_itr_imp >& imp, fs::
             {
                 DWORD error = ::GetLastError();
 
+                if (error == ERROR_NOT_SUPPORTED || error == ERROR_INVALID_PARAMETER)
+                {
+                    // Fall back to file_directory_information_format
+                    filesystem::detail::atomic_store_relaxed(g_extra_data_format, file_directory_information_format);
+                    return dir_itr_create(imp, dir, opts, params, first_filename, sf, symlink_sf);
```
iskiselev, line 886: Attempt to goto into last branch as fallback failed, most probably extra_data / dir_itr_extra_size already has some incorrect data here, have not looked deep.

PR #296, opened 2023-12-05
fix the windows function call, the error logic of the return code

1178264094:
there is a bug fo the return code handle


Lastique:
Thanks.

PR #303, opened 2024-01-14
179 skip fchmod flag (#1)

dsingh-mw:
Added test for the flag "ignore_attribute_errors"


Lastique:
I don't think there's a good way to test this. Replacing `fchmod` (a) requires relying on specific linker behavior and (b) may affect code other than `boost::filesystem::copy_file`. Neither of these I like.

Issue #321, opened 2024-08-25
Warnings/Erros on Windows


Flamefire:
I get a couple issues when compiling on Windows when using Cygwin, MinGW or clang-cl:

```
libs\filesystem\src\operations.cpp(1594,31): error: comparison of integers of different signs: 'boost::winapi::NTSTATUS_' (aka 'long') and 'DWORD' (aka 'unsigned long') [-Werror,-Wsign-compare]
    if (BOOST_UNLIKELY(status == STATUS_INVALID_PARAMETER && (obj_attrs.Attributes & OBJ_DONT_REPARSE) != 0u))
                       ~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~
```

```
libs/filesystem/src/operations.cpp:1404:175: error: cast between incompatible function types from 'boost::winapi::FARPROC_' {aka 'long long int (*)()'} to 'BOOL (*)(HANDLE, boost::filesystem::detail::file_info_by_handle_class, LPVOID, DWORD)' {aka 'int (*)(void*, boost::filesystem::detail::file_info_by_handle_class, void*, long unsigned int)'} [-Werror=cast-function-type]
         GetFileInformationByHandleEx_t* get_file_information_by_handle_ex = (GetFileInformationByHandleEx_t*)boost::winapi::get_proc_address(h, "GetFileInformationByHandleEx");
                                                                                                                                                                               ^
libs/filesystem/src/operations.cpp:1406:166: error: cast between incompatible function types from 'boost::winapi::FARPROC_' {aka 'long long int (*)()'} to 'BOOL (*)(HANDLE, boost::filesystem::detail::file_info_by_handle_class, LPVOID, DWORD)' {aka 'int (*)(void*, boost::filesystem::detail::file_info_by_handle_class, void*, long unsigned int)'} [-Werror=cast-function-type]
         SetFileInformationByHandle_t* set_file_information_by_handle = (SetFileInformationByHandle_t*)boost::winapi::get_proc_address(h, "SetFileInformationByHandle");
                                                                                                                                                                      ^
[...]

PR #322, opened 2024-08-25
Add CI jobs for MinGW32/64

Flamefire:
Add jobs using MSYS2 MinGW to GHA

Config copied from boost-ci

This is the configuration where I see #321 although it seems to be only triggered by `-Wextra` not `-Wall`: https://godbolt.org/z/3fqjG7zM3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -644,8 +644,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - { sys: MINGW32, compiler: gcc, cxxstd: '03,11,17,20' }
-          - { sys: MINGW64, compiler: gcc, cxxstd: '03,11,17,20' }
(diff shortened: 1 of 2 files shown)


Lastique:
- C++03 is not supported
- no `actions/checkout`, please. It causes too much trouble.
- no `boost-ci`, please. I would like to know what and how I'm testing.
- There already is a [MinGW-w64 job](https://github.com/boostorg/filesystem/blob/5dc58be3cddaa00497f9fd95deb8766bb91e2f3a/.github/workflows/ci.yml#L563-L565), why do we need another one?
- I'm not sure what MINGW32 is, but if this is the legacy MinGW32 (as opposed to MinGW-w64) then it is no longer supported. I think, it's been dropped by Boost.System as well.

Flamefire:
CI is passing now in all configs after unconditionally declaring the struct only available to drivers. Anything else?

Lastique:
Thank you.

PR #327, opened 2024-10-01
Enhance test convertible_to_path_and_strings for all char types

Flamefire:
This extends the test in fbd23ee0e07ff17d486534314f89d1b5fb306824 to include all possible `Source` types accepted by `is_convertible_to_path_source`

See : https://github.com/boostorg/filesystem/blob/fbd23ee0e07ff17d486534314f89d1b5fb306824/include/boost/filesystem/detail/path_traits.hpp#L562-L571

### DIFF
--- a/test/path_test.cpp
+++ b/test/path_test.cpp
@@ -161,15 +161,16 @@ public:
     }
 
     operator fs::path() const { return m_path; }
-    operator const fs::path::value_type*() const
-    {
-#if defined(BOOST_WINDOWS_API)
-        return L"[invalid path]";
-#else
-        return "[invalid path]";
+    operator const char*() const { return "[invalid path]"; }
+    operator const wchar_t*() const { return L"[invalid path]"; }
+    operator std::string() const { return "[invalid path]"; }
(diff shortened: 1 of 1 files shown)


Lastique:
This has no effect. The point of the test is to have an ambiguous conversion to a Source type. It doesn't matter which of the candidate types are supported by the original type.

Flamefire:
> This has no effect. The point of the test is to have an ambiguous conversion to a Source type. It doesn't matter which of the candidate types are supported by the original type.

With this enhancement I just want to make sure the ambiguity resolution is not dependent on e.g. the native char type or any of the other possible string-likes.  
The (test) code is very simple and it fits the class name `convertible_to_path_and_strings`, so what does it cost to include it to avoid future regressions?

Lastique:
> With this enhancement I just want to make sure the ambiguity resolution is not dependent on e.g. the native char type or any of the other possible string-likes.

`path` will always support constructors/operations taking native strings, so the existing conversion operators are enough to introduce ambiguity. I don't see the point of adding more operators, it doesn't add anything.

PR #353, opened 2026-08-03
Fix unprovoked symlink errors in directory_entry

### LINKED ISSUE #352: directory_entry reports errors for dangling symlinks when it shouldn't
`directory_entry::symlink_status(error_code&)` and a number of other member functions report errors for dangling symlinks.

The following test program:

```cpp
#include <boost/detail/lightweight_test_report.hpp>

int test_main(int, char*[])
{
    namespace fs = boost::filesystem;
    using boost::system::error_code;

    fs::path dir = fs::temp_directory_path() / "dirent_bug";
    fs::remove_all(dir);
    fs::create_directories(dir);
    fs::create_symlink("nonexistent", dir / "dangling");

    fs::directory_entry entry(dir / "dangling");
    error_code ec;
    fs::file_type t = entry.symlink_status(ec).type();

    BOOST_TEST_EQ(t, fs::symlink_file);
    BOOST_TEST(!ec);

    fs::remove_all(dir);

    return ::boost::report_errors();
}
```

produces the following output:

```
Clang version 21.0.0 (clang-2100.1.1.101), __GXX_EXPERIMENTAL_CXX0X__ defined
libc++ version 210106
Mac OS
Boost version 1.92.0
Command line: ./libs/filesystem/bug/bin/bug
libs/filesystem/bug/bug.cpp(21): test '!ec' failed in function 'int test_main(int, char **)'
1 error detected.
```

The reason is that `directory_entry::refresh_impl` propagates errors from `stat` calls regardless of where it's called from. In v4, this also affects the constructor overload that takes an error code, since the constructor updates the cache.

stm-ableton:
Closes https://github.com/boostorg/filesystem/issues/352.

Also adds tests exercising `directory_entry`.

### DIFF
--- a/include/boost/filesystem/directory.hpp
+++ b/include/boost/filesystem/directory.hpp
@@ -99,6 +99,15 @@ void recursive_directory_iterator_increment(recursive_directory_iterator& it, sy
 BOOST_FILESYSTEM_DECL
 void recursive_directory_iterator_pop(recursive_directory_iterator& it, system::error_code* ec);
 
+enum class directory_entry_update_mask : unsigned int
+{
+    none = 0u,
+    symlink_status = 1u,
+    status = 1u << 1u
+};
+
+BOOST_BITMASK(directory_entry_update_mask)
+
 } // namespace detail
 
 //--------------------------------------------------------------------------------------//
@@ -515,7 +524,7 @@ public:
         ec.clear();
 
         if (!filesystem::status_known(m_status))
-            refresh_impl(&ec, refresh_mode::follow);
+            refresh_impl(&ec, update_mask::status | update_mask::symlink_status);
         return m_status;
     }
 
@@ -523,7 +532,7 @@ public:
     file_status status() const
     {
         if (!filesystem::status_known(m_status))
-            refresh_impl(refresh_mode::follow);
+            refresh_impl(update_mask::status | update_mask::symlink_status);
         return m_status;
     }
 
@@ -546,7 +555,7 @@ public:
         ec.clear();
 
         if (!filesystem::status_known(m_symlink_status))
-            refresh_impl(&ec, refresh_mode::no_follow);
+            refresh_impl(&ec, update_mask::symlink_status);
         return m_symlink_status;
     }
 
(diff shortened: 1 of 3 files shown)


Lastique:
I'd like to let you know that there is a massive change to library headers coming soon, that will likely make PRs against headers non-applicable. You may want to hold off the work on this.

stm-ableton:
@Lastique Thank you for the heads up. I've converted to draft for now.

stm-ableton:
I've rebased around 2466fc17e8a8 (I assumed this was the change to library headers).

include/boost/filesystem/directory.hpp
```diff
@@ -942,7 +942,17 @@ class directory_entry
...
+    {
+        no_follow,
+        follow,
+        follow_lenient
+    };
+
+    BOOST_FILESYSTEM_DECL void refresh_impl(
+        system::error_code* ec = nullptr,
+        refresh_mode mode = refresh_mode::follow_lenient) const;
+    void refresh_impl(refresh_mode mode) const { refresh_impl(nullptr, mode); }
```
Lastique: I'd rather there was a single `refresh_impl`, with the first argument of `refresh_mode`/`update_mask` (with no default) and the second one of `error_core*` (with `nullptr` default). This follows the existing convention.

include/boost/filesystem/directory.hpp
```diff
@@ -942,7 +942,17 @@ class directory_entry
 
 #if !defined(BOOST_FILESYSTEM_DOXYGEN)
 private:
-    BOOST_FILESYSTEM_DECL void refresh_impl(system::error_code* ec = nullptr) const;
+    enum class refresh_mode
+    {
+        no_follow,
+        follow,
+        follow_lenient
```
Lastique: I'm not sure `lenient` is needed. In C++ standard, `std::filesystem::directory_entry` constructor and `assign` are specified to call `refresh`, i.e. those APIs are expected to fail if `status` fails. Other APIs that rely on `status`, obviously, should also fail if `status` fails. The only APIs that only require symlink status are `symlink_status` and `is_symlink`, and those APIs may not update the cached status.

I think, a better way to do this is change `refresh_mode` to `update_mask`, which would be a bit mask of which statuses to update (`update_status`, `update_symlink_status`). `symlink_status` and `is_symlink` would only specify `update_symlink_status` and every other API would specify both.

Lastique (changes requested):

stm-ableton:
@Lastique Thanks. I pushed some fixups, in separate commits to make them hopefully easier to review.

Lastique:
Thanks.

